PR #6, opened 2019-07-25
Updated solution for problem 1

gauravj09:


ashmeetdeol19:
Well Done in this Solution.
Also Submit Solution for Problem 2.

PR #17, opened 2019-08-21
Answers of Problem 1,2

Arpan61:


Problem2.py
```diff
@@ -0,0 +1,28 @@
...
+        def helper(root1, root2):
+
+            if root1 is None and root2 is None:
+                return True
+            if root1 is None or root2 is None:
+                return False
+            else:
+                return (root1.val == root2.val) and helper(root1.right, root2.left) and helper(root1.left, root2.right)
+
+        helper(root, root)
```
sehaj-grover: return helper(root,root)
make this change and it should work

PR #21, opened 2019-09-11
added both the solutions

NihitVarmaSagi:


CheckSymmetry.java
```diff
@@ -0,0 +1,72 @@
+/**
+ * Recursive solution - The time complexity is O(N) and the space complexity is O(Height) where Height will be the utmost stack size.
```
kaurramanpreet: absolutely correct!!

PR #23, opened 2019-09-12
completed pls review

PATIL1695:


PathSum2.py
```diff
@@ -0,0 +1,23 @@
+"""
+The approach here is to start from root and deduct all the intermediate node values in every path and if
+the leaf node equals to the remaining sum value it means we have a valid path.
```
kaurramanpreet: great explanation!!

PR #35, opened 2019-10-30
solutions trees-3

Arihant1467:


Arihant1467:
Yep I have tried after class. I will soon make a pull request

PR #42, opened 2019-10-31
Accepted on leetcode O(n) solution

sudhasr:


symmetric-tree-iterative and recursive.py
```diff
@@ -0,0 +1,63 @@
...
+#         self.right = None
+
+class Solution(object):
+    def isSymmetric(self, root):
+        """
+        :type root: TreeNode
+        :rtype: bool
+        """
+        def helper(root1, root2):
+            if root1 != None and root2 != None:
```
sehaj-grover: your if condition is incorrect 
You need to check if both are null

sudhasr: Okay. I will do it

PR #70, opened 2020-01-28
Solved only problem-2 

nehabindle:
Ran successfully on Leetcode


SymmetricTree_Solution.java
```diff
@@ -0,0 +1,59 @@
+package SymmetricTree;
+
+//import ZigZagOrderTraversal.Solution.TreeNode;
+
```
neha79091: Please write complexity of the solution

PR #74, opened 2020-02-26
before class solution - please review

pbandyop:
@Lohith-Aswathappa


Lohith-Aswathappa:
Good job Payel!

Submit the other code as well

PR #78, opened 2020-02-28
Trees-3 Completed -- Please Review

AbhiniveshP:
Hi,

I've completed Trees-3 as discussed in class and other methods I thought of (iterative)

Thanks,
Abhinivesh

### DIFF
--- a/PathSum.py
+++ b/PathSum.py
@@ -1,0 +1,140 @@
+'''
+Solution:
+1.  The main idea is to add each node's value based on its parents' values and also append the element to a temporary list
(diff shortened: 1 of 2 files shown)


BhushanVasisht:
Good Job. One suggestion is to please put inline comments about what you are doing. Python is very vague and non intuitive. So, it will help us trace the working of the code better

BhushanVasisht:
Good Job. One suggestion is to please put inline comments about what you are doing. Python is very vague and non intuitive. So, it will help us trace the working of the code better

PR #79, opened 2020-02-28
trees3

Pavanisoma:


Symmetric Tree.java
```diff
@@ -0,0 +1,81 @@
+// Time Complexity :o(n)
+// Space Complexity : worst case o(n) or o(height of tree)
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : RECURSIVE
+ I faced issue in line 27 
+ if(nodeA.val !=nodeB.val || nodeA==null || nodeB==null)
+ If I run above condition it is throwing run time error . 
+ How is above code different from current line 27
```
Lohith-Aswathappa: In this case, you check the condition first instead of checking if the node is null or not. In java, if the first condition is true, java doesn't check  if the nodeA and nodeB are null.  So, now your current node will be null, which will be put in the recursive stack. If you try to use that node & find it's children, you get null pointer error.

Best practice is to check if both nodes are not null and then check the specific condition.

Symmetric Tree.java
```diff
@@ -0,0 +1,81 @@
...
+
+If I change my order of if conditions it is throwing run time error.
+
+        if(a.val!=b.val){
+    return false;
+    }
+if(a ==null && b ==null) continue;
+if(a ==null || b ==null){
+    return false;
+    }
```
Lohith-Aswathappa: Always the base conditions have to be on top, so that if you don't do redundant code execution further. 

I believe this is the reason.

Lohith-Aswathappa:
Good job Pavani!

PR #87, opened 2020-03-08
solved both

sivasaimupparaju:
please review


BhushanVasisht:
Good Job. But one suggestion. Don't use variables with same names...in interview while whiteboard coding u will get confused between variables

PR #90, opened 2020-04-08
adds the code for Tree3

shahakshay11:


symmetric_tree.py
```diff
@@ -0,0 +1,33 @@
+"""
+// Time Complexity : O(d) //depth of tree
```
devashishnyati, line 2: It should be O(n). You'll visit all the nodes.

path_sum2.py
```diff
@@ -0,0 +1,51 @@
+"""
+// Time Complexity : O(d2) //d is depth of the tree (Unsure)
```
devashishnyati, line 2: LC says the time complexity is O(n^2). Please verify in the class.

devashishnyati:
Good job Akshay! 
Nice work with the explanations.

PR #94, opened 2020-04-09
Please review

megrao:


Sum_RootLeaf.py
```diff
@@ -0,0 +1,35 @@
+'''
+Time complexity: O(N^2)? Not sure though
```
devashishnyati, line 2: LC says its O(n^2). Please verify it in the class.

symmetricTree.py
```diff
@@ -0,0 +1,34 @@
+"""
+Recurive solution for symmetric binary tree
+Time complexity: O(n)
+Space complexity: O(1)
```
devashishnyati, line 4: You are missing the recursion stack space.

devashishnyati:
Good job Megha!

PR #95, opened 2020-04-09
Completed Trees-3 : Please review

KBangera:
Will add the iterative solution to the second problem soon

### DIFF
--- a/Path_Sum_II.java
+++ b/Path_Sum_II.java
@@ -1,0 +1,29 @@
+// Time Complexity : O(n)
+// Space Complexity : O(h)
+// Did this code successfully run on Leetcode :Yes
+// Any problem you faced while coding this :
+
+
+// Your code here along with comments explaining your approach
+
+class Solution {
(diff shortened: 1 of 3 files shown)


Path_Sum_II.java
```diff
@@ -0,0 +1,29 @@
...
+        List<Integer> currPath = new ArrayList<>();
+        helper(root,sum,currPath,allPaths);
+        return allPaths;
+    }
+    
+    private void helper(TreeNode root,int sum, List<Integer> currPath, List<List<Integer>> allPaths){
+       if(root==null)return;
+        currPath.add(root.val);
+        if(root.left==null && root.right==null && root.val==sum){
+            allPaths.add(new ArrayList<>(currPath));
```
devashishnyati, line 21: So the leetcode explanation says that the time complexity is O(n^2) and in the comments people say it os O(nlogn). Copying the list is O(m) operation where at max m will be logn. This list can be the height of the tree at max. This is my understanding.

devashishnyati:
Good job Karthik.
Verify the time complexity for path sum.

PR #96, opened 2020-04-09
Completed tree 2

arcsoftech:


pathSum.java
```diff
@@ -0,0 +1,46 @@
+// Time Complexity : O(n+m)  where n are nodes and m is edges
```
devashishnyati, line 1: So the leetcode explanation says that the time complexity is O(n^2) and in the comments people say it os O(nlogn). Copying the list is O(m) operation will be logn. This list can be the height of the tree at max. This is my understanding. Worst case it will O(n^2).

devashishnyati:
Good job Arihant.
Verify the time complexity.

PR #97, opened 2020-04-09
Trees 3

vaidyapawan:
101. Symmetric Tree
113. Path Sum II

### DIFF
--- a/101. Symmetric Tree
+++ b/101. Symmetric Tree
@@ -1,0 +1,34 @@
+Time complexity: O(Max Depth)
+Time complexity: O(1)
+Did this code successfully submitted on leetcode: Yes
+
+/**
+ * Definition for a binary tree node.
+ * public class TreeNode {
+ *     int val;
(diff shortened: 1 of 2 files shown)


101. Symmetric Tree
```diff
@@ -0,0 +1,34 @@
+Time complexity: O(Max Depth)
```
devashishnyati, line 1: It will be O(n)

113. Path Sum II
```diff
@@ -0,0 +1,46 @@
+Time complexity: O(Max depth)
```
devashishnyati, line 1: So the leetcode explanation says that the time complexity is O(n^2) and in the comments people say it os O(nlogn). Copying the list is O(m) operation will be logn. This list can be the height of the tree at max. This is my understanding. Worst case it will O(n^2).

devashishnyati:
Good job Pawan!
Please verify the complexities.

PR #98, opened 2020-04-09
Completed Trees 3 Please Review

kannan-mrugapphan:
passes all test cases on leetcode

### DIFF
--- a/PathSum2.java
+++ b/PathSum2.java
@@ -1,0 +1,45 @@
+// 113.
+/**
+ * Definition for a binary tree node.
+ * public class TreeNode {
+ *     int val;
+ *     TreeNode left;
+ *     TreeNode right;
+ *     TreeNode(int x) { val = x; }
+ * }
+ */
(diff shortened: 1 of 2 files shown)


PathSum2.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Definition for a binary tree node.
+ * public class TreeNode {
+ *     int val;
+ *     TreeNode left;
+ *     TreeNode right;
+ *     TreeNode(int x) { val = x; }
+ * }
+ */
+//time - O(n) - visist all nodes in tree
```
devashishnyati, line 11: So the leetcode explanation says that the time complexity is O(n^2) and in the comments people say it os O(nlogn). Copying the list is O(m) operation will be logn. This list can be the height of the tree at max. This is my understanding. Worst case it will O(n^2).

devashishnyati:
Good job Kannan!
Please verify the time complexity.

PR #100, opened 2020-04-10
Adding code for both problems

RutujaKaushike:


PathSum2.java
```diff
@@ -0,0 +1,80 @@
+// Time Complexity : O(n)
```
devashishnyati, line 1: For this question, I was talking about the O(n^2) time complexity. new ArrayList<>(currPath) I believe is O(n) operation.

devashishnyati:
Good job Rutuja!

PR #103, opened 2020-04-16
Completed Trees3

AroopN:


Problem1.py
```diff
@@ -0,0 +1,33 @@
+#
+# Time Complexity  = O(n)
```
devashishnyati, line 2: LC says the time complexity is O(n^2). For every node you are copying a list. The time to copy is O(n). So it becomes O(n^2) in the worst case.

devashishnyati:
Good job Aroop!
Verify the complexity.

PR #104, opened 2020-04-16
Trees 3 problems solved

DebanjanaDasgupta:


Problem1.java
```diff
@@ -0,0 +1,48 @@
...
+    
+    public void helper(TreeNode root, int target, int currSum, List<Integer> currPath, List<List<Integer>> result){
+        if(root == null){
+            return;
+        }
+        
+        if(root.left == null && root.right == null){
+            if(currSum + root.val == target){         
+                
+                List<Integer> temp = new ArrayList<>(currPath);
```
devashishnyati, line 32: The time to copy this list will be O(n).

Problem1.java
```diff
@@ -0,0 +1,48 @@
+// Time complexity: O(n)
```
devashishnyati, line 1: LC says the time complexity for this is O(n^2). You are copying lists in the end and the time to compy that is O(n). Worst case it will be O(n^2).

devashishnyati:
Good job Debanjana!
Verify the complexities!

PR #110, opened 2020-05-13
Solved problems, please review :)

domzhaomathematics:


tree-3.py
```diff
@@ -0,0 +1,154 @@
...
+                if s==sum:
+                    res.append(elements)
+                return
+            helper(root.left,elements[:],s)
+            helper(root.right,elements[:],s)
+        helper(root)
+        return res
+
+#Problem 1: Path Sum II
+#Time Complexity: O(n)
```
sowmyadvn, line 44: For every leaf node (N/2 total leaf nodes), we copy elements into result arraylist O(N) time. So, in total, time complexity would be O(N^2) here

sowmyadvn:
Nice work Dominic!
Please verify time complexity of Path Sum problem

sowmyadvn:
Nice work Dominic!
Please verify time complexity of Path Sum problem

tree-3.py
domzhaomathematics: I wrongly copy pasted my solutions. These lines reffer to the second solution of Path II. Should be fixed now! Thank you for your time :)

PR #111, opened 2020-05-14
Completed Trees-3: Please review

anikpait:


jaspindersingh83:
Good Job Anik. Nice to see all the different solutions

anikpait:
Jaspinder,  I have a quick question:
If you recall, in Path Sum 2, for the brute force solution, we were creating copies of the ArrayList() using the following syntax recursively. 
List<String> l2 = new ArrayList<String>(l1); where l1 is list 1 and l2 is list 2. 
I read here (https://stackoverflow.com/questions/6536094/java-arraylist-copy) that whenever you copy ArrayList, both the ArrayList pointers (l1 and l2) point to the same reference but I remember that you said we create a separate reference while we copy that makes sense.
Because if we create the same reference, then whatever we do changes in l2 will get reflected in l1. If you see the following example: Shouldn't l1 also have 3 elements since l1 and l2 point to same reference?
```
List<String> l1 = new ArrayList<String>();
l1.add("Hello");
l1.add("World");
List<String> l2 = new ArrayList<String>(l1); //A new ArrayList.
l2.add("Everybody");
The result will be l1 will still have 2 elements and l2 will have 3 elements.
```

anikpait:
@jaspindersingh83

jaspindersingh83:
Assigning a reference and creating a new object are two different things.

jaspindersingh83:
l2=l1 (assigning, reference remains same). This is what the stack overflow link is talking about
l2= new ArrayList<>(l1) (creating new copy of l1 which is a separate memory location and assigning it to l2)

anikpait:
Oh okay, makes sense.thanks!

PR #116, opened 2020-05-15
Completed both - Kindly Review

akshay-iyer:


symmetricTree.cpp
```diff
@@ -0,0 +1,36 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
```
sowmyadvn, line 2: Recursive stack space should also be considered auxiliary. So, space complexity would be O(H) here

sowmyadvn:
Good job Akshay!
Please review inline comment regarding complexity

sowmyadvn:
Good job Akshay!
Please review inline comment regarding complexity

PR #123, opened 2020-05-19
PA46 and PA47 solution - Please review

HimanshuAgarwal93:
PA46 :
// Leetcode 113. Path Sum II
// Time complexity: O(n) // As we are traversing each node only once. I am a bit confused about it.
//Space Complexity : O(n) // n is the no of nodes in tree. Space is used for stack.

PA47 : 
// Leetcode 101. Symmetric Tree
//Time Complexity: O(n) // n is no of nodes in tree 
//Space Complexity:O(n) // Stack will have only nodes equal to the heigt of the tree but in worst case it is equal to n.

### DIFF
--- a/PA46.java
+++ b/PA46.java
@@ -1,0 +1,36 @@
+
+// Leetcode 113. Path Sum II
+// Time complexity: O(n) // As we are traversing each node only once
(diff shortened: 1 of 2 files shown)


iivaishnavii:
Great job with both the solutions,Himanshu!

PR #133, opened 2020-06-17
Symmetric Tree

tramatejaswini:


AbhiniveshP:
For iterative approach, the time complexity might seem to look like O(logN), but you traverse all the nodes, and so it would be O(N). Kindly be careful while analyzing Time and Space complexities for Binary Tree questions.

AbhiniveshP:
For iterative approach, the time complexity might seem to look like O(logN), but you traverse all the nodes, and so it would be O(N). Kindly be careful while analyzing Time and Space complexities for Binary Tree questions.

PR #145, opened 2020-06-18
Trees 3 -Please Review

RamaAncha:


Problem46.java
```diff
@@ -0,0 +1,31 @@
...
+        currSum = currSum+node.val;
+        traversedList.add(node.val);
+        if(node.left==null && node.right==null){
+            if(currSum==sum){
+                resultPaths.add(new ArrayList<Integer>(traversedList));
+            }
+        }
+        preOrderTraversePath( node.left, currSum,traversedList, sum);
+        preOrderTraversePath( node.right, currSum,traversedList, sum);
+        traversedList.remove((Integer)node.val);
```
AroopN, line 28: We need to remove the last element added. So, replace this by :
traversedList.remove(traversedList.size()-1);

AroopN:
Good try, Ramakishore. Please check inline comment to fix the Path Sum II problem.

AroopN:
Good try, Ramakishore. Please check inline comment to fix the Path Sum II problem.

PR #166, opened 2020-06-24
Add problem 1 and problem 2

auluckh23:


AbhiniveshP:
Nice work!!!
Time complexity would be O(n) for both the questions.

AbhiniveshP:
Nice work!!!
Time complexity would be O(n) for both the questions.

PR #168, opened 2020-07-09
Ready for review

pallavipsap:
Some glitch in leetcode while running recursion problems ( calling with .copy ). The code runs, but does not get accepted at times

The above problem is fixed automatically , if the code is run after a long time. I do not exactly know if the code had issues or leet code

### DIFF
--- a/46_113_path_sum_II_final (1).py
+++ b/46_113_path_sum_II_final (1).py
@@ -1,0 +1,163 @@
+# Pallavi Sapale
+
+# to understand backtracking
+# deverything fixed
+# yet to try with iterative approach and DFS , yet to ask where to return in backtrack solution exactly
+# similar to 129. root to leaf
+
+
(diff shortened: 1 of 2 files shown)


46_113_path_sum_II_final (1).py
```diff
@@ -0,0 +1,163 @@
+# Pallavi Sapale
+
```
Lohith-Aswathappa, line 2: Try to code the solution as taught in class. It might be that some functions are not accepted in Leetcode but works if other functions are used. Just give it a try with different approach. Let me know if it doesn't work

Lohith-Aswathappa (commented):
Well done, Pallavi!
Review the inline comment.

Lohith-Aswathappa:
Well done, Pallavi!
Review the inline comment.

PR #175, opened 2020-07-30
Completed Trees-3 || Please Review

soumil100:


Problem1.java
```diff
@@ -0,0 +1,76 @@
+//Time Complexity :O(n*n) = O(n^2)
```
AbhijeetDhillon, line 1: Could you explain the time complexity? According to me, it should be O(n)

AbhijeetDhillon:
Nice work Soumil, check the inline comment and let me know what do you think about it.

AbhijeetDhillon:
Nice work Soumil, check the inline comment and let me know what do you think about it.

Problem1.java
soumil100: For each element of the node, you are creating a copy of the paths which takes O(n) worst-case. That's why n* n. A better answer would be n * h where h is the height of the tree. For every node, we need a copy array of h elements. So Time Complexity will be O(n*h).

soumil100:
Checked on Leet code Premium it states, Time Complexity is O(n^2)

PR #179, opened 2020-07-30
Completed Trees-3

alekhya-12:
-Please Review! Thanks!


samarthjj:
Good work Alekhya, Please do try to include more than one solution if applicable!

samarthjj:
Good work Alekhya, Please do try to include more than one solution if applicable!

PR #180, opened 2020-07-30
Trees 3 . Please review - Completed

firoz1905:


Symmetric Tree
```diff
@@ -1,6 +1,6 @@
 Approach1 : Recursive
 Time Complexity : O(N)
-Space Complexity : O(N)
+Space Complexity : Recursive Call . Depth of the tree. Log(N)
```
Sumesh-H: O(h) height of the tree.

Sumesh-H: return helper(left.left,right.right) and helper(left.right,right.left)*

Sumesh-H:
Good work, Firoz. Kindly check inline comment.

Sumesh-H:
Good work, Firoz. Kindly check inline comment.

PR #246, opened 2020-10-08
Completed Trees - Path Sum II - Please review

Sneha9p:


trees2.java
```diff
@@ -0,0 +1,56 @@
...
+ *         this.val = val;
+ *         this.left = left;
+ *         this.right = right;
+ *     }
+ * }
+ */
+
+
+//Time complextiy = O(N square)
+//Space Complexity =O(N)
```
samarthjj, line 19: Time here should be O(n) and Space O(n^2). Every node is visited only once but a new list on length n is created at n nodes, hence n^2.

samarthjj:
Good work Sneha, Please review the inline comments!
Please also complete the Symmetric Tree Question.

samarthjj:
Good work Sneha, Please review the inline comments!
Please also complete the Symmetric Tree Question.

PR #250, opened 2020-10-08
Trees-3

Hanishsai:
Done with Problem1 and Problem2.

### DIFF
--- a/Problem1.java
+++ b/Problem1.java
@@ -1,0 +1,7 @@
+// Time Complexity :
+// Space Complexity :
+// Did this code successfully run on Leetcode :
+// Any problem you faced while coding this :
+
+
+// Your code here along with comments explaining your approach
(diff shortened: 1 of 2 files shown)


Problem2.java
```diff
@@ -0,0 +1,34 @@
+// Time Complexity : O(N)
+// Space Complexity : O(1)
```
kannan-mrugapphan, line 2: Please consider the call stack in space cost too

Problem1.java
```diff
@@ -0,0 +1,48 @@
+// Time Complexity : O(N*N)
+// Space Complexity : O(1)
```
kannan-mrugapphan, line 2: The space can't be constant as you are passing a new list at each call.

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #382, opened 2021-03-10
Please Review

SriharshaVank:


PathSumII.java
```diff
@@ -0,0 +1,37 @@
+// Time Complexity : O(n) we have to visit every node
+// Space Complexity : O(1) auxilary space - constant
```
nsibal, line 2: That's correct, but recursion uses some memory, doesn't it?

SymmetricTree.java
```diff
@@ -0,0 +1,24 @@
+// Time Complexity : O(n) we have to visit every node
+// Space Complexity : O(1) auxilary space - constant
```
nsibal, line 2: That's correct, but recursion uses some memory, doesn't it?

nsibal (commented):
Read inline comments.

nsibal:
Read inline comments.

PR #394, opened 2021-03-15
Completed Trees-3

SahanaVBhat:


pathsum2.java
```diff
@@ -0,0 +1,39 @@
+// Problem1 (https://leetcode.com/problems/path-sum-ii/)
+// Time Complexity : O(N^2) 
+// Space Complexity : O(N)
```
nsibal, line 3: Try improving this.

symmetricTree.java
```diff
@@ -0,0 +1,16 @@
+// Problem2 (https://leetcode.com/problems/symmetric-tree/)
+// Time Complexity : O(N) 
+// Space Complexity : O(N)
```
nsibal, line 3: SC should be O(h) (where h = height of the tree)

nsibal (commented):
Looks good, read inline comments.

nsibal:
Looks good, read inline comments.

pathsum2.java
```diff
@@ -0,0 +1,39 @@
+// Problem1 (https://leetcode.com/problems/path-sum-ii/)
+// Time Complexity : O(N^2) 
+// Space Complexity : O(N) 
+public class pathsum2 {
```
lyndseyBeil, line 4: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names should start with an upper case letter (NM_CLASS_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-should-start-with-an-upper-case-letter-nm-class-naming-convention).
Class names should be nouns, in mixed case with the first letter of each internal word capitalized. Try to keep your class names simple and descriptive. Use whole words-avoid acronyms and abbreviations (unless the abbreviation is much more widely used than the long form, such as URL or HTML).

PR #577, opened 2021-10-29
Completed Trees-3: Please check

adhish70:


brunda4 (commented):
Please provide the correct SC for path sum 2.

brunda4:
Please provide the correct SC for path sum 2.

PR #828, opened 2022-07-19
Add files via upload

sundeep1299:


ishika2741 (commented):
Good work! Kindly add tc and sc for both problems.

ishika2741:
Good work! Kindly add tc and sc for both problems.

PR #959, opened 2023-01-12
Trees-3

agill2144:


pathSum2/main.go
```diff
@@ -0,0 +1,72 @@
+/*
+    approach 1: inorder traversal with recursion
+    - We will save local state at each node
+        - instead of saving a running sum at each node, we will subtract target from root.val
+        - So target will be maintained at each node ( updated target )
+        - We will also keep track of a list thats maintaining running path ( using go slices )*
+            - slices are pass by value
```
daredevilteja, line 7: Instead of using slices can't you use an array here.

pathSum2/main.go
```diff
@@ -0,0 +1,72 @@
...
+            - slices have len and cap as properties
+            - cap = total size avail
```
daredevilteja, line 9: Try to use the array and bring the TC down.

daredevilteja (changes requested):

daredevilteja:
Do the requested chnages.

daredevilteja:
Please do the requested changes.

PR #991, opened 2023-03-23
Done with problem 1!

Ashmitha-art:
Done with problem 1!


Problem1.java
```diff
@@ -0,0 +1,37 @@
+// Time Complexity : O(h)
```
osandeepreddy: TC is O(n)
SC is O(1)

osandeepreddy:
TC is O(n)
SC is O(1)

PR #1020, opened 2023-06-16
Completed Trees-3

pkhanna88:


Problem_1.py
```diff
@@ -0,0 +1,27 @@
+# Time Complexity: O(nlog(n))
+# Space Complexity: O(n)
+# Did this code successfully run on Leetcode: Yes
```
ishika2741, line 3: Does this solution run on leetcode?

ishika2741 (commented):
Good work! Kindly check path sum solution once.

ishika2741:
Good work! Kindly check path sum solution once.

PR #1122, opened 2023-12-10
Tree3

bhavana-nair:


Nandha2Tech (commented):
Great Work! In PathSum2 problem, if you create deep copies then TC would become O(n * h) (h is the time taken for creating new copies). Kindly make a note of it.

Nandha2Tech:
Great Work! In PathSum2 problem, if you create deep copies then TC would become O(n * h) (h is the time taken for creating new copies). Kindly make a note of it.

PR #1148, opened 2024-01-20
Completed Trees-3

keerthana-vijaykumar:


Problem-1.cpp
```diff
@@ -0,0 +1,48 @@
+// Time Complexity : O(N)
+
+// Space Complexity : O(1)
```
ishika2741, line 3: check space complexity , include under hood stack space as well

Problem-2.cpp
```diff
@@ -0,0 +1,42 @@
+// Time Complexity : O(N)
+
+// Space Complexity : O(1)
```
ishika2741, line 3: Check this once

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1161, opened 2024-02-20
JAVA: Trees-3: Completed: RKM

RohanKMeher:


PathSum2.java
```diff
@@ -0,0 +1,59 @@
+// Time Complexity : O(n)
```
ishika2741, line 1: Kindly check time complexity. Also add solution using backtracking as well as that is the optimized solution

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PathSum2.java
RohanKMeher: done

PR #1223, opened 2024-07-14
Trees-3 done

sarthak310:


Symmetric Tree.cpp
```diff
@@ -0,0 +1,47 @@
+// time: O(n)
+// space: O(1)
```
ishika2741, line 2: recursive stack will be using space

ishika2741 (commented):
Well done!

ishika2741:
Well done!

PR #1239, opened 2024-07-27
completed trees-3

dhikki-study:
#######113. Path Sum II########################################################
// Time Complexity : n
// Space Complexity : O(n)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
Following approach is the backtracking approach where I have taken l1 is intermediate list and result as final list, we keep appending nodes to l1 and if any result doesn't match with target we backtrack and remove the element

          
        

######101. Symmetric Tree##################################################
// Time Complexity : n
// Space Complexity : O(n)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : NA

// Your code here along with comments explaining your approach in three sentences only
We have used a BFS approach here, where when we reach to level 2, we take left child of left and right child of right and before inserting we compare, if it fails than return false or else we put it and than for l=next level we compare other childs

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,122 @@
+#######113. Path Sum II#################################################################################################################
+// Time Complexity : n
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : no problem
+
+// Your code here along with comments explaining your approach in three sentences only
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #1307, opened 2025-01-20
Completed both the problems

Erlybird:


Path Sum 2.java
```diff
@@ -0,0 +1,35 @@
+//TC: O(N)
+//SC: O(logN) recursive stack  and path
```
ishika2741, line 2: Please check time complexity

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1472, opened 2025-07-14
Trees-3

thakuryash060:


super30admin:
Strengths:
- Both solutions are correct and handle edge cases properly.
- The use of DFS for pathSum and BFS for isSymmetric is appropriate and efficient.
- Code is clean, readable, and follows good practices.

Areas for Improvement:
- In Problem1.java, consider moving the leaf condition check right after updating `currSum` and `path` to make the logic more intuitive. Currently, it checks after the left DFS call, which might confuse readers.
- Add comments to explain the logic, especially for the BFS approach in Problem2.java, to enhance readability for others.

super30admin:
Strengths:
- Good use of recursion in Problem1 and BFS in Problem2
- Clean code structure and proper naming conventions
- Correct implementation of tree symmetry check in Problem2

Areas for Improvement:
- In Problem1, the order of operations is incorrect. The check for leaf node and target sum match should be done immediately after updating the current sum and path, before making recursive calls to left and right children.
- Consider adding comments to explain the logic, especially for Problem2 where the queue operations might not be immediately obvious to readers.
- The path removal operation in Problem1 is correctly placed, but the overall logic needs adjustment to ensure all paths are captured.

Suggested correction for Problem1:
```java
private void dfs(TreeNode root, int currSum, int targetSum, List<Integer> path) {
    if (root == null) return;
    
    path.add(root.val);
    currSum += root.val;
    
    if (root.left == null && root.right == null && currSum == targetSum) {
        result.add(new ArrayList<>(path));
    }
    
    dfs(root.left, currSum, targetSum, path);
    dfs(root.right, currSum, targetSum, path);
    path.remove(path.size() - 1);
}
```

PR #1476, opened 2025-07-20
Trees-3 Done

JPolishetty:


super30admin:
Strengths:
- Both solutions are correct and efficiently implemented.
- The use of backtracking in Problem1 is a good optimization.
- The helper methods in both problems improve modularity and readability.

Areas for Improvement:
- In Problem1, the comment "as list is mutable, add copy of list" could be expanded to explain why this is necessary for correctness.
- In Problem2, while the solution is correct, it might be worth mentioning that an iterative approach (using a queue) could also be used for symmetry checking, which might be more space-efficient for very large trees.

PR #1490, opened 2025-08-17
Janhvi Guha Trees 3

Janhvi2000:


super30admin:
- Strengths:
  - Good use of helper methods to break down the problem.
  - Proper handling of edge cases (null nodes).
  - Clear and concise code with good comments.
- Areas for Improvement:
  - In ex1.py, the variable name 'res' could be more descriptive (e.g., 'current_sum').
  - The docstrings could be more detailed to explain the purpose of each method.
  - While the space complexity analysis is correct, it might be worth mentioning that in ex1.py, the worst-case space complexity could be O(n^2) when considering the storage of all paths in a skewed tree (though this is unlikely in practice).

PR #1506, opened 2025-09-12
Add solutions for Path Sum II and Symmetric Tree

ravithakur2k:
Implemented PathSum2.py to find all root-to-leaf paths with a given sum and SymmetricTree.py to check if a binary tree is symmetric. Both solutions use recursive approaches and include time and space complexity comments.


super30admin:
Strengths:
- Both solutions are correct and well-implemented.
- Good use of recursion and helper methods.
- Proper handling of edge cases (null checks).
- The backtracking approach in PathSum2.py is more efficient than the reference solution's approach of copying paths.

Areas for Improvement:
- The time complexity analysis for PathSum2.py should be corrected to O(n^2) in worst case.
- For PathSum2.py, consider mentioning the backtracking optimization in comments.
- The space complexity for PathSum2.py should account for path storage in addition to recursive stack.

PR #1517, opened 2025-09-17
Add problem 1 and 2

sarvesh-gaurishankar-sawant:


super30admin:
Strengths:
- Good understanding and implementation of tree traversal and backtracking.
- Code is clean, well-commented, and follows good practices.
- Correct time and space complexity analysis in comments.

Areas for Improvement:
- In Problem2.java, the base case for a null root should return true (an empty tree is symmetric).
- Consider using direct boolean returns in Problem2.java instead of a flag for cleaner code and early termination.
- The comment in Problem1.java about time complexity should mention the worst-case O(n^2) due to path copying.

PR #1543, opened 2025-12-27
Solved Trees-3 problems

aahiawasthee:


super30admin:
Strengths:
- Good use of backtracking in path-sum-ii.java to avoid unnecessary path copies.
- Clear and concise logic in both solutions.
- Correct implementation of both problems.

Areas for Improvement:
- Correct the time and space complexity analysis for path-sum-ii.java. The worst-case time complexity is O(n^2) due to path copying, and space complexity is O(n^2) for the same reason.
- The space complexity claim of O(1) for path-sum-ii.java is misleading; it should be O(n^2) in the worst case.
- Consider adding comments in symmetric-tree.java to explain the approach, as done in path-sum-ii.java.

PR #1546, opened 2026-01-03
Trees-3 solutions

kalyan-ng22:


super30admin:
Strengths:
- The solutions are correct and well-implemented.
- The use of backtracking in PathSum2 is a good optimization.
- Code is clean, readable, and well-commented.

Areas for Improvement:
- The time and space complexity analysis for PathSum2 should be corrected to O(n^2) in the worst case, as path copying can take O(n) time per path in skewed trees.
- For SymmetricTree, the solution is optimal, but the student could consider mentioning the iterative approach (using a queue) as an alternative for completeness.

PR #1558, opened 2026-02-03
Adding Trees-3 solutions

subbu4061:


super30admin:
Strengths:
- The solutions for both problems are correct and efficient.
- The code is clean and well-commented, explaining the approach clearly.
- For path-sum-ii, the use of backtracking with a single list is a good practice to avoid creating multiple copies during traversal.

Areas for Improvement:
- In the path-sum-ii solution, the method parameter `currSum` is passed by value (which is fine in Java since it's primitive), but note that the list is passed by reference. The backtracking step is correctly implemented, but it's important to ensure that the list is modified only when necessary. The current implementation is correct.
- However, in the path-sum-ii solution, the helper function is called even when the current node is null. This is acceptable, but it might be slightly inefficient. Alternatively, you could check for null before making the recursive calls, but the current approach is standard.
- For the symmetric-tree solution, the for-loop inside the while loop is unnecessary because you are polling two nodes at a time. Instead, you can simply poll two nodes in each iteration without the for-loop. The current code uses a for-loop that runs for the entire current level, but then you are polling two nodes at a time. This works, but it might be clearer to avoid the for-loop and just process pairs until the queue is empty. Alternatively, you can process level by level, but the current implementation is correct.

Suggestions:
- In path-sum-ii, you can avoid passing the current sum as a parameter and instead calculate it by subtracting the node value from the target as you go down. This might make the code more intuitive for some, but the current approach is fine.
- For symmetric-tree, consider removing the for-loop since it doesn't add any benefit. You can simply poll two nodes at a time without tracking the level size. However, the current code is correct.

Overall, both solutions are well-written and efficient.

PR #1561, opened 2026-02-09
Complete Trees-3

paridhimalviya:
Problem1 (https://leetcode.com/problems/path-sum-ii/)

Problem2 (https://leetcode.com/problems/symmetric-tree/)

### DIFF
--- a/PathSumTwo.swift
+++ b/PathSumTwo.swift
@@ -1,0 +1,129 @@
+//
+//  RootToLeafSumPathTwo.swift
+//  DSA-Practice
+//
+//  Created by Paridhi Malviya on 1/27/26.
+//
+
+class RootToLeafSumPathTwo {
+    
+    init() {
+        
+    }
+    /*
+     In worst case, h will be n. left skewed tree. O(n2)
+     2.Space complexity - O(n *h) - everytime, creating a new list while deep copying.
+     
+     Backtrack - To be back to the previous state. Whatever we did before the recursion call, we have to remove it. To backtrack the action that we took before recursion. time complexity - O(n), space compelxity - O(h)
+     */
+    //using global result variable
+    
+    var result = [[Int]]()
+    func pathSum(_ root: TreeNode?, _ targetSum: Int) -> [[Int]] {
+        var path: [Int] = []
+        getTargetSumPaths(root, targetSum: targetSum, path: path, currentSum: 0)
+        return result
+    }
+    
+    func getTargetSumPaths(_ root: TreeNode?, targetSum: Int, path: [Int], currentSum: Int) {
+
+       //base
+       guard let root = root else {
+            return
+       }
+
+       //logic
+       let currSum  = currentSum + root.val
+
+       //action
+       var p = path
+       p.append(root.val)
+       if (root.left == nil && root.right == nil) {
+            if (currSum == targetSum) {
+                result.append(p)
+            }
+            return
+       }
+
+       //recurse
+       getTargetSumPaths(root.left, targetSum: targetSum, path: p, currentSum: currSum)
+       getTargetSumPaths(root.right, targetSum: targetSum, path: p, currentSum: currSum)
+
+        //backtrack - not needed because list is value type in Swift, not reference type
+    }
+
+    //MARK: without using global result variable
+    /*
+     time complexityt - O(n * h), in worst case scenario - n can become n - O(n^2)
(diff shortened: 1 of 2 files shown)


super30admin:
Strengths:
- You have implemented multiple approaches, showing a good understanding of different strategies.
- Your code is well-commented, which helps in understanding your thought process.
- You correctly identified the inefficiency of copying the path array and attempted to optimize with backtracking.

Areas for Improvement:
1. The backtracking method should not use a global `result` variable. Instead, you should pass the result as an inout parameter or return it. However, since you are modifying the path in place, it is better to use a global result or pass it by reference.
2. In the backtracking method, you need to ensure that after recursive calls, the path is backtracked by removing the last element. Your current implementation does not do this correctly because you are using `var path = path` which creates a local copy due to value semantics. In Swift, arrays are value types, so when you use `inout`, you are passing a reference to the same array. However, your recursive calls should share the same array instance. But note: your function signature uses `inout`, but you are creating a local variable `var path = path` which defeats the purpose. You should use the `inout` parameter directly without redeclaring it.
3. Remove the print statements in the backtracking method for production code.
4. Avoid including unrelated code (like SymmetricTree) in the same file for clarity.
5. Consider using a helper function that uses backtracking without global variables. For example, you can define a helper that takes `result` as an inout parameter.

Correct backtracking implementation in Swift would look like:
```swift
func pathSum(_ root: TreeNode?, _ targetSum: Int) -> [[Int]] {
    var result = [[Int]]()
    var path = [Int]()
    helper(root, targetSum, 0, &path, &result)
    return result
}

func helper(_ root: TreeNode?, _ targetSum: Int, _ currentSum: Int, _ path: inout [Int], _ result: inout [[Int]]) {
    guard let root = root else { return }
    let newSum = currentSum + root.val
    path.append(root.val)
    if root.left == nil && root.right == nil && newSum == targetSum {
        result.append(path)
    }
    helper(root.left, targetSum, newSum, &path, &result)
    helper(root.right, targetSum, newSum, &path, &result)
    path.removeLast()
}
```
This way, you avoid global variables and efficiently backtrack.

6. Your second method (without global variable) that returns the result is correct but inefficient due to copying. It is acceptable for small trees but not optimal.

super30admin:
Strengths:
- You have explored multiple approaches (global variable, returning results, backtracking), showing a good depth of understanding.
- Your code includes comments explaining time and space complexity, which is excellent for self-review.
- You correctly identify that value-type arrays in Swift lead to copying overhead and that backtracking can improve efficiency.

Areas for Improvement:
1. **Global Variable**: Avoid using global variables for storing results. Instead, pass a mutable reference to the result array through parameters or use a helper function that returns the result. This makes the code more reusable and thread-safe.
2. **Efficiency**: The first method with array copying is correct but inefficient. For large trees, this could be slow. Consider using backtracking with a mutable array that you modify in place.
3. **Backtracking Implementation**: Your backtracking method has several issues:
   - The `inout` parameter `path` is used, but you create a local copy `var path = path` at the start, which defeats the purpose of `inout`. You should use the same mutable array throughout.
   - The backtracking step (`removeLast`) should be done after both recursive calls to ensure that the array is restored to its state before the current node was added.
   - The method does not correctly accumulate results because it uses the global `result` and the recursive calls might not share the array correctly.
4. **Code Organization**: The file contains unrelated code (symmetric tree implementations). It's better to separate different problems into different files for clarity.
5. **Swift Best Practices**: In Swift, using `inout` for recursion can be tricky. Consider defining a helper function that takes a mutable array as a parameter without `inout` by using a class or by using a method that mutates a local array and passes it down.

Suggested backtracking approach without global variable:
```swift
func pathSum(_ root: TreeNode?, _ targetSum: Int) -> [[Int]] {
    var result = [[Int]]()
    var path = [Int]()
    helper(root, targetSum, 0, &path, &result)
    return result
}

private func helper(_ node: TreeNode?, _ target: Int, _ currentSum: Int, _ path: inout [Int], _ result: inout [[Int]]) {
    guard let node = node else { return }
    let newSum = currentSum + node.val
    path.append(node.val)
    if node.left == nil && node.right == nil && newSum == target {
        result.append(path) // Append a copy of the current path
    }
    helper(node.left, target, newSum, &path, &result)
    helper(node.right, target, newSum, &path, &result)
    path.removeLast() // Backtrack
}
```
This approach efficiently uses backtracking to avoid unnecessary array copies.

PR #1563, opened 2026-02-26
Complete Trees-3

PrakarshKamal:


super30admin:
Your solution for "PathSumII" is correct and efficient. Well done! You have correctly implemented DFS with backtracking to collect all root-to-leaf paths that sum to the target. The code is clean and easy to understand.

One minor point: in the `helper` method, you are passing the current sum and the path as parameters. However, note that the path is being modified by adding and then removing elements (backtracking). This is a good practice to avoid creating multiple copies of the path during recursion.

Also, you are creating a new ArrayList when adding to the result (`ans.add(new ArrayList<>(path))`), which is necessary to avoid the list being modified later.

However, I noticed that your file also contains a solution for "SymmetricTree". While this is not incorrect, it is generally better to have one class per file, especially in Java. If this is part of a larger project, consider separating the solutions into their own files.

Overall, great job!

PR #1569, opened 2026-03-08
working solution

avcode3:


super30admin:
Your solution for problem1.py is correct and efficient. You have correctly implemented the DFS traversal with backtracking. The code is clean and easy to understand. Well done!

One minor point: In the helper function, you check `if root.left is None and root.right is None` to identify a leaf node. This is correct. However, you can also write it as `if not root.left and not root.right` which is more concise but equally clear.

Also, note that the reference solution uses a parameter `currNum` to pass the current sum, while you use `curr_sum`. Both are fine. Your solution uses a mutable list for the path, which is modified in place and backtracked by popping. This is efficient and avoids copying the entire path at each node (only copies when a valid path is found).

Overall, excellent job!

PR #1571, opened 2026-03-15
Add implementation for leetcode problems 113, 101

rishigoswamy:


super30admin:
### Root-to-leaf Sum II (leetcode_101.py)
It appears there has been a misunderstanding. The problem you were asked to solve is "Root-to-leaf Sum II", which requires returning all root-to-leaf paths where the sum of node values equals the targetSum. However, your solution is for "Symmetric Tree", which checks if a binary tree is symmetric.

To solve the correct problem, you need to traverse the tree (using DFS or BFS) and keep track of the current path and current sum. When you reach a leaf node, check if the current sum equals the targetSum. If it does, add the current path to the result.

Here are some pointers for the correct solution:
- Use a recursive DFS approach to traverse the tree, passing the current path and current sum.
- When you visit a node, add its value to the current sum and push the value to the current path.
- At a leaf node (both children are null), check if the current sum equals the targetSum. If yes, add the current path to the result.
- Remember to backtrack: remove the current node from the path when returning from the recursive call (if using a mutable list that is passed by reference).
- Alternatively, you can pass a new copy of the path at each recursive call to avoid backtracking, but this may be less efficient in terms of space.

Your current code does not address the problem at all. Please review the problem statement again and implement the correct algorithm.

VERDICT: NEEDS_IMPROVEMENT

---

### Mirror Image of Itself Tree (leetcode_113.py)
Dear student,

I notice that your solution is for a different problem (Path Sum II) than the one described (Symmetric Tree). This is a critical issue. Please ensure that you are solving the correct problem. For the Symmetric Tree problem, you need to check if the tree is symmetric around its center, which requires comparing the left and right subtrees.

Regarding your Path Sum II solution:
- Your primary DFS approach with backtracking (using `array.append` and `array.pop`) is correct and efficient. It avoids recalculating the sum by maintaining `currSum`, which is good.
- However, you have included several alternative methods that are either inefficient or incorrect. For example, using `sum(array)` at each node would lead to O(n^2) time complexity, which is not optimal. Also, creating copies of the array at each node (as in the third method) would consume excessive memory.
- It's good to explore different approaches, but in a final solution, you should only include the most efficient and clean method. The commented code might be for your own learning, but for submission, keep only the best version.
- For the Symmetric Tree problem, you would need to implement a recursive or iterative solution that checks if the left subtree is a mirror of the right subtree. The reference solution uses an iterative BFS approach with a queue to compare nodes in pairs.

Please double-check the problem statement next time. For Symmetric Tree, you can start with the following recursive approach:
```
def isSymmetric(root):
    if not root:
        return True
    return isMirror(root.left, root.right)

def isMirror(left, right):
    if not left and not right:
        return True
    if not left or not right:
        return False
    return (left.val == right.val and 
            isMirror(left.left, right.right) and 
            isMirror(left.right, right.left))
```

Or an iterative approach using a queue similar to the reference solution.

VERDICT: NEEDS_IMPROVEMENT

